PR #209, opened 2021-06-06
Configurable blame virtual text position

andersevenrud:
This PR adds a configuration option to adjust the `virt_text_pos` for the git blame virtual text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ require('gitsigns').setup {
     interval = 1000
   },
   current_line_blame = false,
+  current_line_blame_position = 'eol',
   sign_priority = 6,
   update_debounce = 100,
   status_formatter = nil, -- Use default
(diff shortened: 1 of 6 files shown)


andersevenrud:
Oh, silly me. This is written in Teal.... fixing that.

andersevenrud:
That should be taken care of now.

lewis6991:
LGTM. Thanks a lot.

andersevenrud:
Apparently, the right align option for virtual text does not respect the actual text of the buffer, so if the blame line gets really long it will conceal the text :/ I didn't catch this at first because I only show first name and date with a custom formatter.

Here's an example of what the problem:

2021-06-07-012304_3840x2160_scrot

I'm in the process of finding a solution for this though, but that will probably make this option obsolete.

So maybe revert this commit ?! Sorry about that.

andersevenrud:
I'll probably open an issue in neovim about this, because it would be nice to have an option that actually aligns to the right but also crops the text so that there's no overlap.

lewis6991:
Yeah maybe it is worth reverting.

PR #240, opened 2021-07-03
show status message when navigating hunks

### LINKED ISSUE #237: Display warning message when trying to go to next/previous hunk if there isn't one
I think it would be a good idea to display a warning message on the status line saying that there are no more hunks after/before the last/first one (of course, only when the wrap option is set to false). Other popular git sign packages like gitgutter have this functionality. What are your thoughts on this?

zapling:
Closes #237

### DIFF
--- a/types/types.d.tl
+++ b/types/types.d.tl
@@ -191,6 +191,7 @@ global record vim
     diffopt: string
     wrapscan: boolean
     updatetime: number
+    shortmess: string
   end
 
 
--- a/doc/gitsigns.txt
+++ b/doc/gitsigns.txt
@@ -113,6 +113,10 @@ next_hunk({opts})                                           *gitsigns.next_hunk(
                             • {wrap}: (boolean)
                               Whether to loop around file or not. Defaults
                               to the value 'wrapscan'
+                            • {navigation_message}: (boolean)
+                              Whether to show navigation messages or not.
+                              Looks at 'shortmess' for default behaviour.
+ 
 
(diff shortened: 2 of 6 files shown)


mroavi:
Hey @zapling ! Thanks for implementing this! I have a small suggestion. Would it be possible to print "No more hunks" in red color?

zapling:
@mroavi  I added the highlight group `WarningMsg` so it now looks like this

Screenshot_2021-07-03_13-53-51

mroavi:
Awesome. Thanks. I just tried it out and it is working great.

mroavi:
One last thing that might be useful is an option to enable it or disable it. I can imagine that there are people out there who might not want this.

zapling:
Sounds responsible

lewis6991:
Is there not a default vim option similar to `wrapscan` that we can piggy back off for default behaviour? For displaying results of the `/` command, if `'shortmess'` contains `S`, then results are not shown.

lewis6991:
LGTM

zapling:
I'll just update the docs

types/types.d.tl
```diff
@@ -191,6 +191,7 @@ global record vim
     diffopt: string
     wrapscan: boolean
     updatetime: number
+    shortmess: string
```
zapling, line 194: I just added this value manually, is there another way?

lewis6991: Nope, It's all manual. We just add to it as we need.

PR #295, opened 2021-08-06
remove use_decoration_api from README.md

laantorchaweb:
I got this warning that `use_decoration_api` is now removed so I checked here and the option was still on the docs.

Relevant commit on removing the option: https://github.com/lewis6991/gitsigns.nvim/commit/27e1e95e1aa8d4ec0aabbf5148013f58a1bc4b40

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,6 @@ require('gitsigns').setup {
   update_debounce = 100,
   status_formatter = nil, -- Use default
   word_diff = false,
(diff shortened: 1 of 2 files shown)


gegoune:
There also is https://github.com/lewis6991/gitsigns.nvim/blob/27e1e95e1aa8d4ec0aabbf5148013f58a1bc4b40/doc/gitsigns.txt#L73.

PR #327, opened 2021-08-22
Fix warning text to state current_line_blame_position is deprecated, …

SahilMadan:
…not current_line_blame_opts


lewis6991:
Fixed in b3466a1187f7212277b904f27f2bed08182c52a6

PR #509, opened 2022-04-03
Fix `indent_heuristic` data type in the documentation

EPadronU:
The documentation states `indent_heuristic` is a string when it should be `boolean`.


lewis6991 (changes requested):
You need to modify `teal/gitsigns/config.tl` and re-gen the docs. Should be explained in CONTRIBUTING.md

PR #514, opened 2022-04-10
Make signs thicker

David-Kunz:
Hi 👋 ,

To make the signs more readable, one can use the unicode characters [left half block](https://www.htmlsymbols.xyz/unicode/U+258C), [upper one eighth block](https://www.htmlsymbols.xyz/unicode/U+2594) and [lower one eighth block](https://www.htmlsymbols.xyz/unicode/U+2581).

It looks like this:
Screen Shot 2022-04-10 at 11 08 53

Compare it to the default:
Screen Shot 2022-04-10 at 11 09 09

Maybe you want to change the default?

Thanks for looking into it and best regards,
David

### DIFF
--- a/lua/gitsigns/config.lua
+++ b/lua/gitsigns/config.lua
@@ -115,10 +115,10 @@ M.schema = {
       type = 'table',
       deep_extend = true,
       default = {
-         add = { hl = 'GitSignsAdd', text = '│', numhl = 'GitSignsAddNr', linehl = 'GitSignsAddLn' },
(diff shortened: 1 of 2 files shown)


lewis6991:
I think I would rather change the default to `┃` which is a little thicker than the current default.

David-Kunz:
Thanks for the feedback, I switch it to the mentioned character, it looks like this:

Screen Shot 2022-04-15 at 17 23 16

lewis6991:
Thanks

PR #839, opened 2023-07-16
Add GitHub integration

alexmarucci:
Following up on the original PR https://github.com/lewis6991/gitsigns.nvim/pull/623

- Added `github_blame` and github formatter config
- Added PR info in line blame
- Added PR number in line blame popup
- If no PR is found, falls back to the std git blame

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,11 @@ require('gitsigns').setup {
     virt_text = true,
     virt_text_pos = 'eol', -- 'eol' | 'overlay' | 'right_align'
     delay = 1000,
+    github_blame = false,
     ignore_whitespace = false,
   },
   current_line_blame_formatter = '<author>, <author_time:%Y-%m-%d> - <summary>',
+  current_line_blame_formatter_gh = ' <author>, <mergedAt:%Y-%m-%d>, PR: #<number> • <title>',
   sign_priority = 6,
   update_debounce = 100,
   status_formatter = nil, -- Use default
--- a/lua/gitsigns/config.lua
+++ b/lua/gitsigns/config.lua
@@ -432,15 +432,15 @@ M.schema = {
   count_chars = {
     type = 'table',
     default = {
-      [1] = '1', -- '₁',
-      [2] = '2', -- '₂',
-      [3] = '3', -- '₃',
-      [4] = '4', -- '₄',
-      [5] = '5', -- '₅',
-      [6] = '6', -- '₆',
-      [7] = '7', -- '₇',
-      [8] = '8', -- '₈',
-      [9] = '9', -- '₉',
+      [1] = '1',   -- '₁',
+      [2] = '2',   -- '₂',
+      [3] = '3',   -- '₃',
+      [4] = '4',   -- '₄',
+      [5] = '5',   -- '₅',
+      [6] = '6',   -- '₆',
+      [7] = '7',   -- '₇',
+      [8] = '8',   -- '₈',
+      [9] = '9',   -- '₉',
       ['+'] = '>', -- '₊',
     },
     description = [[
@@ -674,7 +674,6 @@ M.schema = {
   current_line_blame_formatter_gh = {
     type = { 'string', 'function' },
     default = ' <author>, <mergedAt>, PR: #<number> • <title>',
-    --- TODO: confirm <author_time> in github
     description = [[
       String or function used to format the virtual text of
       |gitsigns-config-current_line_blame| when github is used.
--- a/lua/gitsigns/config.lua
+++ b/lua/gitsigns/config.lua
@@ -46,6 +46,7 @@ end
 --- @field delay integer
 --- @field ignore_whitespace boolean
 --- @field virt_text_priority integer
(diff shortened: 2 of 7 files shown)


alexmarucci:
Hi @lewis6991, I hope PRs are welcomed. This is a feature I wanted for a while now, so I thought giving it a go.

Let me know your thoughts, also I might need a hand with the failing tests. Unsure why they're failing to be fair.

Thank you

README.md
```diff
@@ -75,9 +75,11 @@ require('gitsigns').setup {
     virt_text = true,
     virt_text_pos = 'eol', -- 'eol' | 'overlay' | 'right_align'
     delay = 1000,
+    github_blame = false,
```
alexmarucci, line 78: @lewis6991 should this be called something like `use_github` or something like that instead?

lua/gitsigns/config.lua
```diff
@@ -669,6 +671,26 @@ M.schema = {
...
+      |gitsigns-config-current_line_blame| when github is used.
+
+      Note: |gitsigns-config-current_line_blame_opts-github_blame| must be active.
+
+      When a string, accepts the following format specifiers in addition to the defaults:
+
+          • `<number>`
+          • `<author>`
+          • `<mergedAt>` or `<mergedAt:FORMAT>`
+          • `<title>`
```
alexmarucci, line 688: As an improvement, we could expand this to the full list of fields that `gh` gives us instead? only nested fields are not supported yet

lua/gitsigns/current_line_blame.lua
```diff
@@ -136,6 +137,18 @@ local function run_blame(bufnr, lnum, opts)
...
   local bcache = cache[bufnr]
   result = bcache.git_obj:run_blame(buftext, lnum, opts.ignore_whitespace)
+
+  if result and opts.github_blame then
+    local last_pr = gh.get_last_associated_pr(bcache.git_obj.repo.toplevel, result.sha);
+
+    if last_pr then
+      result = last_pr;
+      -- the parser does not support accessing keys like <author.name>
+      result.author = last_pr.author.name
```
alexmarucci, line 147: `<author.name>` maybe that's something we could add? so we could just query the GH endpoint and pass down  all the available fields

lewis6991:
Hi. PRs are definitely welcome, however I have a relatively high bar for new features (just to be warned).

I'll have a fair amount of feedback to give, but I'll need to test this first.

Like everyone, I'm pretty busy, spread quite thin across several projects and only limited time to work on open source stuff so it might be a while until I can look at this properly.

One thing I'll say now is that this feature must not affect the experience of the plugin on non-github repos and therefore should purely be additive.

I'm also not so sure about adding another formatter field to the config. The formatter (string version) was added without the consideration of optional fields, whereas a single function can handle all cases.

The plugin must also not try to call `gh` if it is not installed, not just simply ignore the error.

Other than that, this looks like a good start.

alexmarucci:
> Hi. PRs are definitely welcome, however I have a relatively high bar for new features (just to be warned).
> 
> I'll have a fair amount of feedback to give, but I'll need to test this first.
> 
> Like everyone, I'm pretty busy, spread quite thin across several projects and only limited time to work on open source stuff so it might be a while until I can look at this properly.
> 
> One thing I'll say now is that this feature must not affect the experience of the plugin on non-github repos and therefore should purely be additive.
> 
> I'm also not so sure about adding another formatter field to the config. The formatter (string version) was added without the consideration of optional fields, whereas a single function can handle all cases.
> 
> The plugin must also not try to call `gh` if it is not installed, not just simply ignore the error.
> 
> Other than that, this looks like a good start. 👍

Nice, thank you for your feedback. That's totally fine, and I think I got enough to make few adjustments already 👍 I will see if I get some time to look into the failing tests.

In general, I see this as a starting point to make some progress on this feature following some directions from your side.

Speak soon 👋

mechanicles:
These changes seem very useful for people who work with Git and Github. Most of the time, we do want to see the git commit message for the current line of code, and we also want to see its related Github PR number so that we can refer it to see all the changes for that code. If this PR solves this issue, I would really appreciate getting it merged.

mechanicles:
@alexmarucci Great work. Actually I tried your branch in my neovim settings and found that it works for some cases and for some cases it does not show Github PR number.

With PR number:
Screenshot 2023-11-22 at 8 22 12 PM

With No PR number:
Screenshot 2023-11-22 at 8 22 43 PM

mechanicles:
> These changes seem very useful for people who work with Git and Github. Most of the time, we do want to see the git commit message for the current line of code, and we also want to see its related Github PR number so that we can refer it to see all the changes for that code. If this PR solves this issue, I would really appreciate getting it merged.

Not sure why my reply got marked as spam but I really feel this PR is really useful.

lewis6991:
Commenting a PR is useful **is** spam.

mechanicles:
> Commenting a PR is useful **is** spam.

Ahh.. ok. Sorry for that. Did not know about it.

PR #840, opened 2023-07-17
fix: no commit message

Aryonal:
Fix no commit message.

image

After fix

image


Aryonal:
Hi @lewis6991 

I hope you're doing well! I happened to find out about this issue from my local, and I tried this fix and it's working for me, so I create this PR. Since this is my first PR here, sorry if I'm not doing it the right way.

Please let me know if there are any concerns and feedback, and thank you for your wonderful repo :)

PR #942, opened 2024-01-27
Update the README to match the new default attached_to_untracked setting

hankertrix:


hankertrix:
I also updated the gitsigns.txt file.

PR #1099, opened 2024-07-25
feat(blame): add verbose render style

iamxiaojianzheng:
image

image

### DIFF
--- a/lua/gitsigns/actions.lua
+++ b/lua/gitsigns/actions.lua
@@ -1058,9 +1058,21 @@ end
 ---
 --- Attributes: ~
 ---     {async}
-M.blame = async.create(0, function()
-  return require('gitsigns.blame').blame()
-end)
+---
+--- Parameters: ~
+---     See |gitsigns.blame()|.
+M.blame = function(print_type)
+  if print_type ~= nil then
+    print_type = tostring(print_type)
+  end
+
+  -- set default to simple
+  if print_type ~= 'simple' and print_type ~= 'verbose' then
+    print_type = 'simple'
+  end
+
+  require('gitsigns.blame').blame(print_type)
+end
 
 --- @param bcache Gitsigns.CacheEntry
 --- @param base string?
--- a/lua/gitsigns/blame.lua
+++ b/lua/gitsigns/blame.lua
@@ -59,40 +59,51 @@ local chars = {
 
 local M = {}
 
+--- @param print_type 'simple' | 'verbose'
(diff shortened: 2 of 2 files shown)


lewis6991:
What is the point of this?

iamxiaojianzheng:
> What is the point of this?


This might be a better read than the original default. After all, the previous style, there are mismatches. 

I don't think it's very intuitive.

iamxiaojianzheng:
This is an alternative and does not change the default behavior. 

You can switch to this style using parameter control. 

`Gitsigns blame 'verbose'`

iamxiaojianzheng:
The difference looks something like this

image image

iamxiaojianzheng:
I refer to the display effect of `idea`, and I feel that this will indeed be more intuitive.

image

The original way, I think it's a bit of a waste of effort to guess the range of gaps.

lewis6991:
Sorry but I don't want to maintain this, so I think I'll pass.

The current behaviour matches what Github does.

iamxiaojianzheng:
These thick dotted lines feel very affecting reading. After using it for a while, it was a little hard to accept.

image

lewis6991:
Those are defined by your config. Try:

```lua
vim.opt.fillchars:append { diff = ' ' }
```

iamxiaojianzheng:
> Those are defined by your config. Try:
> 
> 
> 
> ```lua
> 
> vim.opt.fillchars:append { diff = ' ' }
> 
> ```

The current display problem is not the character, but the height. No matter what character is changed, even if it is blank, the final effect will interfere with reading.

iamxiaojianzheng:
> Sorry but I don't want to maintain this, so I think I'll pass.
> 
> 
> 
> The current behaviour matches what Github does.
> 
> 
> 
> image
> 
> 

The final effect is also very different from that in the picture.

lewis6991:
One is in a terminal, and the other is in a gui. What changes would you suggest to make it better?

iamxiaojianzheng:
> Looks close enough to me:看起来离我很近：
> 
> image
> 
> One is in a terminal, and the other is in a gui. What changes would you suggest to make it better?一个在终端中，另一个在 gui 中。你建议做出哪些改变来让它变得更好？



> Looks close enough to me:
> 
> image
> 
> One is in a terminal, and the other is in a gui. What changes would you suggest to make it better?

Haha, it doesn't matter what style it is. Just get used to it.

PR #1162, opened 2025-01-06
feat: file blame options

GrzegorzKozub:
Default is kept:

image

The new options and highlights let us match the current editor color theme:

image


GrzegorzKozub:
Alright, no problem.

PR #1267, opened 2025-04-21
Small documentation updates

omrisarig13:
Firstly, thanks for a great plugin.

I've started using it recently, and while copying the example keymaps from the README file noticed that one of the used function is marked as deprecated.

I've removed this function from the README example and fixed a small typo in the contributing guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ make gen_help
 
 ## Testsuite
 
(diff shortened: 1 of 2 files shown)


omrisarig13:
@lewis6991 There is still the first commit, with the deprecated example from README (which I think is the more relevant change), you can decide whether you want to get it in or not.

PR #1414, opened 2025-09-30
feat(cache): add support for fetching line ranges

PeterCardenas:
I have this util for prefetching blame to make the popup show up faster, but it's not performant when fetching blame line individually, even when running them concurrently. Thus I changed the API to allow for fetching blame for a line range instead of just one line. I thought this functionality would be generally useful which is why I'm pushing upstream here, and also am thinking this prefetch utility in a cleaner form could also be upstreamed.

```lua
vim.api.nvim_create_autocmd({ 'BufEnter', 'CursorMoved', 'WinScrolled' }, {
  group = vim.api.nvim_create_augroup('gitsigns-prefetch-blame', { clear = true }),
  callback = function(opts)
    local bufnr = opts.buf
    local cache_entry = require('gitsigns.cache').cache[bufnr]
    if not cache_entry then
      return
    end
    local buf_line_count = vim.api.nvim_buf_line_count(bufnr)
    if blame_fetch_map[bufnr] then
      return
    end
    blame_fetch_map[bufnr] = true
    local config = require('gitsigns.config').config
    local gitsigns_async = require('gitsigns.async')
    ---@type fun(lnum: integer|[integer, integer], on_done: fun():nil): nil
    local prefetch_blame = gitsigns_async.create(
      2,
      ---@async
      ---@param lnum integer|[integer, integer]
      ---@param on_done fun():nil
      ---@return Gitsigns.BlameInfo?
      function(lnum, on_done)
        -- Defers error notification to hover keymap
        pcall(function() ---@async
          return cache_entry:get_blame(lnum, config.current_line_blame_opts)
        end)
        on_done()
      end
    )
    local cursor_lnum = vim.api.nvim_win_get_cursor(0)[1]
    local start_lnum = math.max(1, vim.fn.line('w0') - 10, cursor_lnum - 20)
    local end_lnum = math.min(buf_line_count, vim.fn.line('w$') + 10, cursor_lnum + 20)
    prefetch_blame({ start_lnum, end_lnum }, function()
      blame_fetch_map[bufnr] = false
    end)
  end,
})
```

### DIFF
--- a/lua/gitsigns/cache.lua
+++ b/lua/gitsigns/cache.lua
@@ -98,7 +98,7 @@ local BLAME_THRESHOLD_LEN = 10000
 
 --- @async
 --- @private
---- @param lnum? integer
+--- @param lnum? integer|[integer, integer]
 --- @param opts? Gitsigns.BlameOpts
 --- @return table<integer,Gitsigns.BlameInfo?>
 --- @return table<string,Gitsigns.CommitInfo?>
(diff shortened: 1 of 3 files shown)


lua/gitsigns/cache.lua
```diff
@@ -157,29 +157,55 @@ end
...
+    end
+  else
+    blame_valid = self:blame_valid(lnum)
+  end
+  if not blame or not blame_valid then
     self:wait_for_hunks()
     blame = blame or { entries = {} }
     local Hunks = require('gitsigns.hunks')
-    if lnum and Hunks.find_hunk(lnum, self.hunks) then
+    local has_hunks = true
```
lewis6991, line 180: Should this init to `false`?

lua/gitsigns/cache.lua
```diff
@@ -157,29 +157,55 @@ end
...
+    if lnum then
+      local start_lnum = type(lnum) == 'table' and lnum[1] or lnum
+      local end_lnum = type(lnum) == 'table' and lnum[2] or lnum
+      for curr_lnum = start_lnum, end_lnum do
+        has_hunks = not not Hunks.find_hunk(curr_lnum, self.hunks)
```
lewis6991, line 185: Should this early return, otherwise won't this just be the value of the last iteration?

PeterCardenas: yea ur right, had this as a while loop before then blindly changed every while to a for when cleaning things up :/

lewis6991: This is a little hard to follow. I guess this is saying that we run blame if there is at least 1 line that isn't in a hunk.

PR #1419, opened 2025-10-04
fix: workaround emmylua param infer for async.create wrappers

mochaaP:
before:
before

after:
after

also updated docgen to not reflect the typings in the generated docs.


lewis6991:
Don't think this is needed anymore.

image

PR #1439, opened 2025-12-10
feat: enhance status formatting with highlight colors

lopi-py:
## Before
Blank statusline
image

## After
Colorfull statusline
image

Neovim offers something similar with `vim.diagnostic.status()`

### DIFF
--- a/doc/gitsigns.txt
+++ b/doc/gitsigns.txt
@@ -804,9 +804,15 @@ status_formatter                            *gitsigns-config-status_formatter*
         function(status)
           local added, changed, removed = status.added, status.changed, status.removed
           local status_txt = {}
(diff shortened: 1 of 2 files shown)


lopi-py:
Hello @lewis6991, can you explain why was this reverted in 718e1eda17ebab87f686af5dcc7684b204041eef? It works with every colorscheme I tried, even the default one in Neovim:

Statusline:
image

Signcolumn:
image

It doesn't override background highlights. In any case, colorschemes should not set a background color for these highlights and instead just set the foreground

lewis6991:
When I tried it with my statusline it caused the background for that section to change.

If you want this, you need to add it manually to your config using the dict variant.

PR #1448, opened 2026-01-01
fix(word_diff): no "No newline at end of file" shown in popup

phanen:
Also add `GitSignsNoEOLPreivew` hlgroup.
* https://github.com/tpope/vim-fugitive/blob/2019e0e4139390f485a024d7a2411218b004a5b3/syntax/fugitive.vim#L30
* https://github.com/neovim/neovim/blob/35a7642647858f7b4ddc204ee869c399b678e7e8/runtime/syntax/diff.vim#L379

Test
```sh
mkdir ./repro && cd ./repro && git init
echo a >a && git add . && git commit -m 'this file has newline'
printf a >a && nvim --clean --cmd 'set rtp^=..' a +'lua vim.defer_fn(function() vim.cmd[[Gitsigns preview_hunk]] end, 100)'
```

### DIFF
--- a/doc/gitsigns.txt
+++ b/doc/gitsigns.txt
@@ -1273,6 +1273,11 @@ GitSignsDeletePreview
         Used for deleted lines in previews.
 
         Fallbacks: `GitGutterDeleteLine`, `SignifyLineDelete`, `DiffDelete`
+                                                        *hl-GitSignsNoEOLPreview*
+GitSignsNoEOLPreview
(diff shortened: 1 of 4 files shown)


lewis6991:
Can you add a test?

phanen:
Added.

lewis6991:
`GitSignsNoEOLPreivew` should be `GitSignsNoEOLPreview`

PR #1470, opened 2026-01-20
Calculate line length correct with inlay hints

svenbergner:
The line length wasn't calculated correct with activated inlay hints.
With blame line align right could it happen that the blame text covered the source code.


lewis6991:
Thanks

svenbergner:
Your welcome. Thanks for accepting my PR.

Issue #1135, opened 2024-10-20
Add filetype to popups


gabs712:
### Description

gitsigns.preview_hunk and gitsigns.blame_line popups currently doesn't have filetypes.
I suggest adding one for each, so them can be referenced.

### Neovim version

v0.11.0

### Operating system and version

Ubuntu 24.04.1

### Actual behavior

Running `:set ft?` inside of popups outputs a null value.

### Minimal config

```Lua
require('gitsigns').setup()
```

### Steps to reproduce

1. Trigger gitsigns.blame_line or gitsigns.preview_hunk
2. Trigger it again to enter the window
3. Run `:set ft?`
4. There isn't a filetype defined

Picalines:
+1

See https://github.com/mfussenegger/nvim-dap/issues/415 on why it's useful

PR #1474, opened 2026-01-29
chore: add buffer filetype to preview_hunk

camiloaromero23:
Addresses https://github.com/lewis6991/gitsigns.nvim/issues/1135
image
image

 Open to suggestions

### DIFF
--- a/lua/gitsigns/actions/preview.lua
+++ b/lua/gitsigns/actions/preview.lua
@@ -199,6 +199,15 @@ local function noautocmd(f)
   end
 end
 
+local function withautocmd(f)
+  return function()
+    local ei = vim.o.eventignore
+    vim.o.eventignore = ''
+    f()
+    vim.o.eventignore = ei
+  end
+end
+
 --- Preview the hunk at the cursor position in a floating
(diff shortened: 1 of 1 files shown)


phanen:
I think https://github.com/lewis6991/gitsigns.nvim/issues/1135 is about a event/FileType-event specific to popup, but not for syntax hl. 

For syntax hl, trigger `FileType` event is not a good idea. It may also trigger lsp server or other plugins (which also listen on FileType, scan your ftplugin/ in rtp, although this is negligible). Also it would be better to exclude the diff related signs like `-`/`+` when parse the text, I guess it may not work well with some language, more or less.

IMHO set something like `ft=gitsigns_popup` is enough for user, then they can use the event for their own syntax hl, e.g. they can borrow utils in fancy diff in snacks https://github.com/folke/snacks.nvim/commit/22eea90a9548e692c80a20740934720e6d095be1.

phanen:
FYI. Make a [plugin](https://github.com/phanen/blame_hl.nvim) just now. It monkey patch `Popup.update/create` to get preview_buf/preview_winid.


image

Issue #1349, opened 2025-06-24
Gitsigns fails to attach in msys2


goulf-3m:
### Description

The debug messages showed that extra '\n' are present in git commands.

This fixes. Please apply or do similar newline striping!

```
diff --git a/lua/gitsigns/git/repo.lua b/lua/gitsigns/git/repo.lua
index 3ed6196..cdad8cb 100644
--- a/lua/gitsigns/git/repo.lua
+++ b/lua/gitsigns/git/repo.lua
@@ -168,6 +168,9 @@ local function normalize_path(path)
     -- through cygpath
     --- @type string
     path = async.await(3, system, { 'cygpath', '-aw', path }).stdout
+    if path:sub(-1) == '\n' then
+      path = path:sub(1, -2)
+    end
   end
   return path
 end
```

### Neovim version

Not relevant

### Operating system and version

Windows 11

### Actual behavior

Gitsigns failed to attach

### Minimal config

```Lua
for name, url in pairs{
  gitsigns = 'https://github.com/lewis6991/gitsigns.nvim',
  -- ADD OTHER PLUGINS _NECESSARY_ TO REPRODUCE THE ISSUE
} do
  local install_path = vim.fn.fnamemodify('gitsigns_issue/'..name, ':p')
  if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system { 'git', 'clone', '--depth=1', url, install_path }
  end
  vim.opt.runtimepath:append(install_path)
end

require('gitsigns').setup{
  debug_mode = true, -- You must add this to enable debug messages
  -- ADD GITSIGNS CONFIG THAT IS _NECESSARY_ FOR REPRODUCING THE ISSUE
}

-- ADD INIT.LUA SETTINGS THAT IS _NECESSARY_ FOR REPRODUCING THE ISSUE
```

### Steps to reproduce

Run nvim in MSYS2

### Gitsigns debug messages

```text

```

### Gitsigns cache

```lua

```

goulf-3m:
The above is for Gitsigns 1.0.2.

For the master branch (1b0350a), it's complaining:
```
'C:\Users\goulf\Desktop\configs' is outside worktree '/c/Users/goulf/Desktop/configs'
```

goulf-3m:
This BUG is still around. The debug message says:

```
 'C:\Users\xxx\Desktop\configs' is outside worktree '/c/Users/xxx/Desktop/configs'
```

[...]

PR #1500, opened 2026-03-24
Fix accessible paths being ommited from normalization on windows systems

Dioarya:
This problem affects me and gould-3m has found a solution, I hope there's no side effects.
This fixes when vim can see the path exists, but incorrectly omits it from being normalized.
I'm on windows MSYS2. I think this also fixes things on Cygwin, but not sure.


Dioarya:
Fixes #1349. Credit to gould-3m for this fix.

lewis6991:
@Dioarya can you check https://github.com/lewis6991/gitsigns.nvim/pull/1501

lewis6991:
Fixed in #1501

PR #1515, opened 2026-03-29
feat(heatmap): create configurable highlight group for temperature colors

dimtion:
So I was seeing that the current commit heatmap colors doesn't allow much configuration.

The current implementation (inspired by fugitive I see?)  is using  tri-color pattern Blue / Red / Yellow for indicating the age of the commit. On some color schemes, this makes the red pop more than the yellow: giving the impression that red commits are newer than yellow ones.

I'm proposing here a couple of improvement:
1. Using a single bi-color scheme (cold-hot) and do a smoothstep interpolation between then two. Which is generally considered a better practice to show a scale for a variable.
2. Let the user or theme define  and override the `GitSignsColorTemp` highlights
3. Fallback to `Normal` (bg = cold, fg = hot), which in my opinion is more subtle for an age indicator.

I'm happy to discuss the color/theme decisions since those can be subjective. Let me know what you think!

<details>
<summary>See screenshots</summary>
Screenshot 2026-03-29 at 16 24 53
Screenshot 2026-03-29 at 16 25 22
Screenshot 2026-03-29 at 16 27 24
Screenshot 2026-03-29 at 16 26 05
Screenshot 2026-03-29 at 16 26 32
</details>

### DIFF
--- a/doc/gitsigns.txt
+++ b/doc/gitsigns.txt
@@ -1354,6 +1354,11 @@ GitSignsVirtLnum
     Used for line numbers in inline hunks previews.
 
     Fallbacks: `GitSignsDeleteVirtLn`
+                                                        *hl-GitSignsColorTemp*
+GitSignsColorTemp
+    Heatmap color: fg = hot (newer changes), bg = cold (older changes)
+
+    Fallbacks: `Normal`
 
 ==============================================================================
 COMMAND                                                      *gitsigns-command*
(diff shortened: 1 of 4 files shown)


dimtion:
Hi @lewis6991 I was wondering if you would be interested in such change, and if so, if PR requires more work, either on the premise (bi-color), the way the color scheme is defined or on the implementation.

lewis6991:
Haven't had time to review.

I want to keep a 2 stop approach.

PR #1559, opened 2026-06-18
feat(blame): Add option to show blamed commit in the existing window

drakari7:
In the blame buffer, when you press <CR> on a commit, you get a pmenu with different options. This MR adds an aditional option to see the commit in the current window. This avoids having to do a :tabnew which also ends up creating a [No Name] buffer.


drakari7:
@lewis6991 This is a purely additive functional change. If you're not opposed to it, I would love to see it merged! Thanks for taking a look.

PR #1568, opened 2026-07-20
Allow 2 count characters and show count character options in README

nullromo:
## [feat] Allow 2 columns of count chars

Only 2 columns of sign chars are allowed by nvim. It makes sense to allow users to utilize both of these. Generally, for a diff where the count matters, the very first sign char (like `+`) isn't that important since the whole section will be marked with sign chars. For this reason, it makes sense to allow the use of that first char on the first line for another purpose.

For example, the UI might look like this normally if there are 3 lines changed

```
+3
+
+
```

I would argue that the first `+` is not strictly needed since the `3` is there and will also be subject to the same highlight group.

If there are 13 lines changed, the UI looks like this

```
+>
+
+
+
+
+
+
+
+
+
+
+
+
```

Notice that there is not enough space for the number `13`, meaning the best we can do is `+>`.

This pull request allows 2-digit `count_chars` entries, which end up looking like this

```
13
+
+
+
+
+
+
+
+
+
+
+
+
```

It's a little bit better since it allows for **proper counts on diffs of up to 99 lines**, not just 9 lines.

## [docs] Add `show_count` and `count_chars` options to README

 These options are documented in the docs, but are not in the README. Most users will look for configuration and settings in the README for a quick glance at what's available. It's nice to mention the settings there, even if they are already explained in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,21 +156,22 @@ with most of the default settings:
 ```lua
 require('gitsigns').setup {
   signs = {
-    add          = { text = '┃' },
-    change       = { text = '┃' },
-    delete       = { text = '_' },
-    topdelete    = { text = '‾' },
-    changedelete = { text = '~' },
-    untracked    = { text = '┆' },
+    add          = { text = '┃', show_count = false },
+    change       = { text = '┃', show_count = false },
+    delete       = { text = '_', show_count = false },
+    topdelete    = { text = '‾', show_count = false },
+    changedelete = { text = '~', show_count = false },
+    untracked    = { text = '┆', show_count = false },
   },
   signs_staged = {
-    add          = { text = '┃' },
-    change       = { text = '┃' },
-    delete       = { text = '_' },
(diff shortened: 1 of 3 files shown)


nullromo:
Here is a screenshot of what it looks like

image

And here's the settings I used for this:

```lua
local superscripts = { ['0'] = '⁰', ['1'] = '¹', ['2'] = '²', ['3'] = '³', ['4'] = '⁴', ['5'] = '⁵', ['6'] = '⁶', ['7'] = '⁷', ['8'] = '⁸', ['9'] = '⁹' }
local countChars = { ['+'] = '*>' }
for index = 1, 99 do
    countChars[index] = tostring(index):gsub('.', superscripts)
end
require('gitsigns').setup({
    -- ...
    count_chars = countChars,
})
```

nullromo:
@lewis6991 Any chance you could take a look at this PR?

PR #1574, opened 2026-08-13
fix(git): resolve the working tree of a linked worktree gitdir

Hnasar:
When `GIT_DIR` is set but `GIT_WORK_TREE` is not, `get_info` derives
the working tree as `dirname(gitdir)`. For a linked `git worktree` the
`gitdir` is `.git/worktrees`.

https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

This mean gitsigns thinks that all opened files are not part of the
working tree, and so gitsigns never attaches.

Read `<gitdir>/gitdir` which contains the path to the working tree
`.git` file (which means the parent of that is the `worktree`'s working
tree).

Absent that file the previous `dirname` is still correct, so plain
repositories and the redirected-`.git` case above are unaffected.

I found this because git exports `GIT_DIR` to `!shell` aliases, so an
alias that opens vim from a linked worktree causes gitsigns to not
attach.

Repro with:

    GIT_DIR=/main/.git/worktrees/w nvim /w/some/tracked/file

Before, `:lua =require('gitsigns.cache').cache[0]` is nil; after, the
buffer attaches and `toplevel` is `/w`.

Additional fixes:
* Also includes a port of the fix from https://github.com/lewis6991/nvim-test/pull/17
* Pass pathspecs relative to the worktree
* Pathspecs are now separated with --, since relative paths can clash with revision names like HEAD 
Windows-only, and it survived the first fix. MSYS git rewrites absolute arguments (D:/a/x → /d/a/x); for a linked worktree we also pass --work-tree (only detached repos do), so the rewritten pathspec stops matching it and git rejects the file as outside the repository it just named:

And when testing, these harness bugs caused spurious errors:
- Scratch cleanup before any Nvim session exists. cleanup_scratch_root runs on busted's suite start, before the first session, and reached is_win() — an RPC into that session. assert(session) fired and the suite died during setup, running zero tests.
- Symlinked checkouts. scratch_root came from PJ_ROOT verbatim while Nvim and git report resolved paths, so every path assertion failed when the checkout sat under a symlinked home.

### DIFF
--- a/lua/gitsigns/git/repo.lua
+++ b/lua/gitsigns/git/repo.lua
@@ -590,6 +590,25 @@ local function normalize_path(path)
   return vim.fs.normalize(path)
 end
 
+--- Working tree for a gitdir, when it can't be discovered from the cwd.
+--- @async
+--- @param gitdir string
+--- @return string
+local function get_worktree(gitdir)
+  -- A linked worktree's GIT_DIR may be set to `.git/worktrees/<name>/`.
+  -- The gitdir file contains the path to the working tree.
+  -- https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables
+  local dotgit = read_first_line(Path.join(gitdir, 'gitdir'))
+  if dotgit then
+    -- `worktree.useRelativePaths` stores this relative to gitdir.
+    if not Path.is_abs(dotgit) then
+      dotgit = Path.join(gitdir, dotgit)
+    end
+    return vim.fs.dirname(assert(normalize_path(dotgit)))
+  end
+  return vim.fs.dirname(gitdir)
+end
+
 --- @async
 --- @param gitdir string
 --- @param head_str string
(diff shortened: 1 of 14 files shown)


lua/gitsigns/git/repo.lua
```diff
@@ -590,6 +590,21 @@ local function normalize_path(path)
...
+--- @async
+--- @param gitdir string
+--- @return string
+local function get_worktree(gitdir)
+  -- A linked worktree's GIT_DIR may be set to `.git/worktrees/<name>/`.
+  -- The gitdir file contains the path to the working tree.
+  -- https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables
+  local dotgit = read_first_line(Path.join(gitdir, 'gitdir'))
+  if dotgit then
+    return vim.fs.dirname(assert(normalize_path(dotgit)))
```
lewis6991, line 656: Git can store this path relatively when worktree.useRelativePaths or git worktree add --relative-paths is used. In that case dotgit looks like ../../../../linked/.git, so this resolves it relative to Neovim's cwd rather than gitdir. Could we join relative values to gitdir before normalizing?

Hnasar: oh TIL about this new option, thanks & fixed.
(Sorry for the debug spam, wish I had a windows box to test with)

Hnasar:
(wow what an ordeal to get to the bottom of that async hang. Will clean this up and un-draft it)

lewis6991:
ah dam, any feedback to make it better/easier to debug?

